PR #2, opened 2025-04-01
Add Polaris benchmarks

pingtimeout:
This pull request introduces a comprehensive set of benchmarks to Polaris. The current set includes:

* A benchmark that populates an empty Polaris server with a dataset that have predefined attributes
* A benchmark that issues only read queries over that dataset
* A benchmark that issues read and write queries (entity updates) over that dataset, with a configurable read/write ratio
* A benchmark that creates commits against existing tables

A documentation if provided in the README.md file, including examples of the different datasets that can be generated. The datasets are procedural, which means that given the same input parameters, the same datasets will be generated, thus enabling reproducible benchmarks.

Note that there is one big change compared to the initial PR that was opened against `apache/polaris`: the benchmark configuration is now handled via a configuration file (see the README.md file).  Using environment variables was brittle and could not be easily documented.


pingtimeout:
@jbonofre just a heads up, I might need some help to get the license check and header check added to that project.  I did not implement it as I thought, maybe we will have a parent project in this repository too, just like in `apache/polaris`.  Let me know how you want to proceed.

benchmarks/README.md
```diff
@@ -0,0 +1,242 @@
...
+  namespace-depth = 4                            # Depth of the namespace tree
+  tables-per-namespace = 5                       # Tables per namespace
+  views-per-namespace = 3                        # Views per namespace
+  columns-per-table = 10                         # Columns per table
+  columns-per-view = 10                          # Columns per view
+  default-base-location = "file:///tmp/polaris"  # Base location for datasets
+  namespace-properties = 10                      # Number of properties to add to each namespace
+  table-properties = 10                          # Number of properties to add to each table
+  view-properties = 10                           # Number of properties to add to each view
+  max-tables = -1                                # Maximum tables (-1 for unlimited)
```
RussellSpitzer, line 65: How do these two properites (max-tables/ max-views) interact with the specifications above?

benchmarks/README.md
```diff
@@ -0,0 +1,242 @@
...
+
+The diagrams below describe the data sets that are used in benchmarks. Note that the benchmark dataset may not cover all Polaris features.
+
+## Generation rules
+
+The dataset has a tree shape. At the root of the tree is a Polaris realm that must exist before the dataset is created.
+
+An arbitrary number of catalogs can be created under the realm. However, only the first catalog (`C_0`) is used for the rest of the dataset.
+
+The namespaces part of the dataset is a complete `N`-ary tree. That is, it starts with a root namespace (`NS_0`) and then, each namespace contains exactly `0` or `N` children namespaces. The width as well as the depth of the namespaces tree are configurable. The total number of namespaces can easily be calculated with the following formulae, where `N` is the tree arity and `D` is the total tree depth, including the root:
```
RussellSpitzer, line 161: just a note, I believe you described N as width previously and not arity? Or am I mistaking this? Just wondering if we should change this to also say width for consistency

benchmarks/build.gradle.kts
```diff
@@ -0,0 +1,29 @@
...
+    scala
+    id("io.gatling.gradle") version "3.13.5.2"
+    id("com.diffplug.spotless") version "7.0.2"
+}
+
+description = "Polaris Iceberg REST API performance tests"
+
+tasks.withType<ScalaCompile> {
+    scalaCompileOptions.forkOptions.apply {
+        jvmArgs = listOf("-Xss100m") // Scala compiler may require a larger stack size when compiling Gatling simulations
```
RussellSpitzer, line 11: :eyes:

benchmarks/src/gatling/scala/org/apache/polaris/benchmarks/NAryTreeBuilder.scala
```diff
@@ -0,0 +1,95 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.polaris.benchmarks
+
+case class NAryTreeBuilder(nsWidth: Int, nsDepth: Int) {
```
RussellSpitzer, line 22: May be helpful to add in some details on the implementation of this class. Looks to me like this is essentially a class which maps a range of ordinal values into a tree structure. Is that correct?

benchmarks/src/gatling/scala/org/apache/polaris/benchmarks/actions/AuthenticationActions.scala
```diff
@@ -0,0 +1,100 @@
...
+/**
+ * Actions for performance testing authentication operations. This class provides methods to
+ * authenticate and manage access tokens for API requests.
+ *
+ * @param cp Connection parameters containing client credentials
+ * @param accessToken Reference to the authentication token shared across actions
+ * @param maxRetries Maximum number of retry attempts for failed operations
+ * @param retryableHttpCodes HTTP status codes that should trigger a retry
+ */
+case class AuthenticationActions(
```
RussellSpitzer, line 44: Is this part of the benchmarks listed in the config? I don't remember seeing a doc in the readme for this

benchmarks/src/gatling/scala/org/apache/polaris/benchmarks/actions/CatalogActions.scala
```diff
@@ -0,0 +1,135 @@
...
+  /**
+   * Creates a new Iceberg catalog with FILE storage type. The catalog is created as an INTERNAL
+   * type with a name and a default base location that are defined in the [[CatalogActions.feeder]].
+   * This represents the fundamental operation of establishing a new catalog in an Iceberg
+   * deployment.
+   *
+   * There is no limit to the number of users that can create catalogs concurrently.
+   */
+  val createCatalog: ChainBuilder =
+    retryOnHttpStatus(maxRetries, retryableHttpCodes, "Create catalog")(
```
RussellSpitzer, line 79: Looks good for now, but I was wondering if this should be using a Polaris client rather than doing String based querying? Probably not important because I assume we are going to keep these apis static but I generally am afraid of string constants.

benchmarks/README.md
```diff
@@ -0,0 +1,242 @@
...
+
+The benchmarks use synthetic procedural datasets that are generated deterministically at runtime. This means that given the same input parameters, the exact same dataset structure will always be generated. This approach allows generating large volumes of test data without having to store it, while ensuring reproducible benchmark results across different runs.
+
+The diagrams below describe the data sets that are used in benchmarks. Note that the benchmark dataset may not cover all Polaris features.
+
+## Generation rules
+
+The dataset has a tree shape. At the root of the tree is a Polaris realm that must exist before the dataset is created.
+
+An arbitrary number of catalogs can be created under the realm. However, only the first catalog (`C_0`) is used for the rest of the dataset.
```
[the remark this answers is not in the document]
pingtimeout: I am not sure what you mean here?

benchmarks/src/gatling/scala/org/apache/polaris/benchmarks/NAryTreeBuilder.scala
pingtimeout: You are absolutely correct.  I can add some javadoc, yes.  But it will be mostly a repeat of the readme without the charts that give visual examples.

benchmarks/src/gatling/scala/org/apache/polaris/benchmarks/actions/AuthenticationActions.scala
pingtimeout: I assume you are talking about the `*Actions.scala` classes.  These classes regroup the elements that are needed to perform certain actions based on either a feature (e.g. authenticating) or an entity (e.g. tables).

They are not listed in the README, indeed, as they are an implementation detail.  This is also where the configuration is consumed, not defined.

benchmarks/src/gatling/scala/org/apache/polaris/benchmarks/actions/CatalogActions.scala
pingtimeout: It is a good question.  In my experience, Gatling benchmarks should have little, if any, external dependencies, so that they are easy to read.  The risk is for Gatling benchmarks to become only readable/editable with a fully fledged IDE, instead of a simple text editor or even Github UI.

We _could_ extract those String constants into some sort of a Polaris client.  But that would be it.  The connection logic and all that would make a complete Polaris **client** would not be there, as they are Gatling specific.  So I am not sure it would be net positive.

benchmarks/README.md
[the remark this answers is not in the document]
RussellSpitzer: Ah nvm, I missed the previous sentence and was like where did realm come from :) My bad

benchmarks/src/gatling/scala/org/apache/polaris/benchmarks/actions/AuthenticationActions.scala
RussellSpitzer: I mean specifically the Authentication Actions, I didn't understand where that fit in the Write/Read benchmark

pingtimeout: Ok I see.  The explanation is that before any request can be issued, the user must first authenticate against the OAuth endpoint.  This class takes care of that.  It can feel a little bit over-engineered as, at this time, authentication is performed only once at the very beginning of the benchmark.  Example from the `CreateTreeDatasetConcurrent` simulation.

```
    authenticate
      .inject(atOnceUsers(1))
      .andThen(createCatalogs.inject(atOnceUsers(50)))
      .andThen(
        ..
      )
```

But the OAuth token is only valid for 1h.  Which means that any benchmark that lasts more than that will fail.  I have separate modifications (not published yet) that renew this token periodically that make use of the `AuthenticationActions` more.  Does that help clarify?

RussellSpitzer: I believe we should be using the credential refresh hooks now? I haven't followed this enough in the OSS Iceberg a discussion but there should be auto-refresh hooks now but that may just be for the Iceberg client?

benchmarks/src/gatling/scala/org/apache/polaris/benchmarks/util/CircularIterator.scala
```diff
@@ -0,0 +1,52 @@
...
+    currentIterator.next()
+  }
+}
+
+class BufferedRandomIterator[T](underlying: CircularIterator[T], bufferSize: Int)
+    extends Iterator[T] {
+  private var buffer: Iterator[T] = populateAndShuffle()
+
+  private def populateAndShuffle(): Iterator[T] =
+    Random.shuffle((1 to bufferSize).map(_ => underlying.next()).toList).iterator
```
RussellSpitzer, line 42: Do we need to seed this? not sure if we were doing determinism for the iterators here as well

RussellSpitzer (approved):
I did a general skim here and everything looks good to me. I won't have time to get through a detailed review in the near future but wanted to give my stamp of approval.

benchmarks/src/gatling/scala/org/apache/polaris/benchmarks/actions/AuthenticationActions.scala
pingtimeout: We could.  But I am not sure what the value would be for OSS Polaris.  I need to research that more.  AFAICT, the OAuth endpoint is deprecated for removal.  Which means there could be a way to bypass authentication entirely later for internal testing purposes.  Alternatively, if it only stays for development purposes, then we don't need to strictly follow the client credentials flow?

benchmarks/build.gradle.kts
```diff
@@ -0,0 +1,29 @@
...
+    }
+}
+
+dependencies {
+    gatling("com.typesafe.play:play-json_2.13:2.9.4")
```
dimas-b, line 16: nit: do we want to use `toml` files like the main Polaris repo?

benchmarks/src/gatling/resources/gatling.conf
```diff
@@ -0,0 +1,10 @@
+gatling {
+  charting {
+    indicators {
+      percentile1 = 25
+      percentile2 = 50
+      percentile3 = 75
```
dimas-b, line 6: IQR :tada: ;)

benchmarks/src/gatling/scala/org/apache/polaris/benchmarks/NAryTreeBuilder.scala
```diff
@@ -0,0 +1,105 @@
...
+    (math.log(numerator) / math.log(nsWidth)).floor.toInt
+  }
+
+  /**
+   * Calculate the total number of nodes in a complete n-ary tree.
+   *
+   * @return The total number of nodes in the tree.
+   */
+  val numberOfNodes: Int = {
+    print(s"Computing number of nodes for n-ary tree with width $nsWidth and depth $nsDepth...")
```
dimas-b, line 71: Why `print` here?

benchmarks/src/gatling/scala/org/apache/polaris/benchmarks/actions/AuthenticationActions.scala
dimas-b: At some point Polaris will support external IdP (AFAIK), so auto-refresh hooks may not cover all cases.

benchmarks/src/gatling/scala/org/apache/polaris/benchmarks/util/CircularIterator.scala
dimas-b: +1 to supporting seeds, but we ca probably add that later.

dimas-b (approved):

benchmarks/build.gradle.kts
ajantha-bhat: I have introduced `toml` in #1, once we merge it we need to rebase this PR and use it.

benchmarks/README.md
pingtimeout: Added some more examples in the reference configuration file as well as clarified the 1-line description in the readme.

benchmarks/build.gradle.kts
pingtimeout: I believe this comment applies for the line just after `play-json` (the `typesafe-config` line).  Play-json is used to parse the payloads returned by Polaris and ensure that maps (e.g. namespace properties) are equal.  Given that there is no order guarantee between properties, a plain string comparison cannot be used.  So play-json has to stay.

We _could_ move from typesafe-config to a `toml` file.  But I would first like to double check that we are talking about the same thing.  Typesafe config was initially preferred as it is already used to configure Gatling and offers the ability to have default parameter values for benchmarks that can then be overridden by users either from the CLI or a separate configuration (HOCON) file.

AFAICT in #1, toml files are used for the Gradle build.  The equivalent of typesafe-config in #1 is picocli, not `toml` files.  And I missing something @ajantha-bhat?

benchmarks/src/gatling/resources/gatling.conf
pingtimeout: Yes that one is specifically for your box plots :-)

benchmarks/src/gatling/scala/org/apache/polaris/benchmarks/util/CircularIterator.scala
pingtimeout: Good point, I had overlooked that.  Fixed with the last push.

benchmarks/src/gatling/scala/org/apache/polaris/benchmarks/NAryTreeBuilder.scala
pingtimeout: Removed

pingtimeout:
I will wait for #1 to be merged as it contains the base build directives that could be leveraged in this PR.  I might have to modify these base Gradle files though, in order to enable Scala projects.  It looks like #1 assumes that all projects under this repository will be Java-based.  To be continued.

benchmarks/build.gradle.kts
dimas-b: I mean `toml` for gradle dependencies, not for benchmark config... sorry about the confusion :sweat_smile:

dimas-b (approved):

benchmarks/README.md
```diff
@@ -0,0 +1,242 @@
...
+  software distributed under the License is distributed on an
+  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  KIND, either express or implied.  See the License for the
+  specific language governing permissions and limitations
+  under the License.
+-->
+
+# Polaris Benchmarks
+
+This repository contains benchmarks for the Polaris service using Gatling.
```
snazy, line 22:
```suggestion
Benchmarks for the Polaris service using Gatling.
```
It's not its own repository

snazy (approved):
LGTM!

Looks like it doesn't conflict with #1, so +1 to merge.

jbonofre (approved):

PR #6, opened 2025-04-16
Remove sequential simulations and add continuous auth to Polaris benchmarks

pingtimeout:
This PR is part of a preliminary work before more benchmarks are added.

* Remove sequential simulations now that the concurrent simulation throughput is configurable (and can be set to 1 until SQL implementation is able to keep up with throughput)
* Configure every workload using the configuration file
* Change authentication logic so that OAuth token is refreshed every minute. This makes it possible to run a benchmark for longer than the default OAuth validity period (1h).  It is useful for use cases like creating very large data sets or running longevity tests.


benchmarks/src/gatling/scala/org/apache/polaris/benchmarks/simulations/CreateTreeDataset.scala
```diff
@@ -64,11 +67,31 @@ class CreateTreeDataset extends Simulation {
...
   // --------------------------------------------------------------------------------
-  val authenticate: ScenarioBuilder = scenario("Authenticate using the OAuth2 REST API endpoint")
-    .feed(authenticationActions.feeder())
-    .exec(authenticationActions.authenticateAndSaveAccessToken)
+  val continuouslyRefreshOauthToken: ScenarioBuilder =
+    scenario("Authenticate every minute using the Iceberg REST API")
+      .asLongAs(_ => shouldRefreshToken.get())(
+        feed(authenticationActions.feeder())
+          .exec(authenticationActions.authenticateAndSaveAccessToken)
+          .pause(1.minute)
```
dimas-b, line 80: nit: Could we make this configurable too?

dimas-b (approved):
LGTM :+1: Thanks, @pingtimeout !

benchmarks/src/gatling/scala/org/apache/polaris/benchmarks/simulations/CreateTreeDataset.scala
pingtimeout: Well, we could but it would be a tricky parameter.

This action is equivalent to an uninterruptible sleep.  So in the worse case, with this 1 minute value, Gatling will continue running for nothing during 60 seconds after the last response from all other queries has been received.

If we make this parameter, I can imagine how people would think that a token needs be refreshed only every 30 minutes.  And now they have to wait up to 30 minutes before the process actually completes.

I will likely give another try at this element in the future.  Like, instead of a 1 minute pause, run a loop of 60 "sleep 1s" actions with the same check `_ => shouldRefreshToken.get()` that is done earlier.  Or something similar.

benchmarks/src/gatling/scala/org/apache/polaris/benchmarks/simulations/ReadTreeDataset.scala
```diff
@@ -61,13 +64,31 @@ class ReadTreeDataset extends Simulation {
...
+        feed(authenticationActions.feeder())
+          .exec(authenticationActions.authenticateAndSaveAccessToken)
+          .pause(1.minute)
+      )
+
+  val waitForAuthentication: ScenarioBuilder =
+    scenario("Wait for the authentication token to be available")
+      .asLongAs(_ => accessToken.get() == null)(
+        pause(1.second)
+      )
```
eric-maynard, line 84: nit:m I guess we don't have a concept of an established scala style, but this formatting looks awkward to me. I'm used to seeing either:
```
.asLongAs(_ => accessToken.get() == null)(pause(1.second))
```

Or preferably:
```
.asLongAs(_ => accessToken.get() == null) {
  pause(1.second)
}
```

eric-maynard (approved):

benchmarks/src/gatling/scala/org/apache/polaris/benchmarks/simulations/CreateTreeDataset.scala
adutra: I agree this is misleading, as you could easily infer that changing to `.pause(30.minutes)` wouldn't have any consequence on the test duration.

Suggestion (but I don't know Gatling very well): instead of continuously polling the atomic integer until it becomes false, couldn't the caller send an interrupt signal to the scenario?

But it's OK imo to investigate improvements later.

adutra (approved):

benchmarks/src/gatling/scala/org/apache/polaris/benchmarks/simulations/CreateTreeDataset.scala
dimas-b: Thanks for the clarification! I'm ok with improving this later.

benchmarks/src/gatling/scala/org/apache/polaris/benchmarks/simulations/ReadTreeDataset.scala
pingtimeout: Given that the parenthesis and curly braces are interchangeable in this case, I just updated all occurrences to the block syntax.  I agree that having the pause on a new indentation level is more readable.

pingtimeout:
Just a heads up, I believe all comments/questions are addressed.  But as I am not a committer, I need need help from a maintainer to merge this PR.

dimas-b (approved):

PR #9, opened 2025-04-21
Benchmarks docs, config and core reorganisation

pingtimeout:
~This PR should only be reviewed after #6 is merged as it includes all the changes from that PR.~ **Edit:** Ready for review now that #6 has been merged.

This PR contains the following changes:

* e3790a2: Cleanup of README so that it is more concise
* dc68d6b and a70f310: Rework of benchmarks parameters so that each benchmark is configured by a dedicated section in the config file
* d0f4e8a: Remove the updates generation parameters so that benchmarks that update the dataset are easier to configure

More on d0f4e8a:

This commit changes the way update requests are generated.  They used to be generated up to a certain number.  Now as many updates as the user wants can be generated, using infinite streams.

This not only simplifies the configuration of update-related benchmarks but also reduce the amount of boilerplate code.

Typically the `BufferedRandomIterator` class was needed to shuffle updates, as otherwise, multiple updates against the same entity would be sent simultaneously.  Now, before an entity is updated again, all the other entities of the dataset need to receive an update as well.  This removes the need for the shuffling and buffering operation.

As a result, there are no random operation that are performed anymore in the simulations.  The seed parameter is not necessary anymore.  It has been removed as it is dead code.

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -23,7 +23,7 @@ Benchmarks for the Polaris service using Gatling.
 
 ## Available Benchmarks
 
-- `org.apache.polaris.benchmarks.simulations.CreateTreeDataset`: creates a test dataset with a specific structure.  It is a write-only workload designed to populate the system for subsequent benchmarks.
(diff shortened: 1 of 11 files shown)


snazy (approved):
LGTM!

PR #11, opened 2025-04-21
Added append only, upsert only modes for polaris-synchronizer, and put modification aware behind opt-in flag.

mansehajsingh:
Currently the polaris-synchronizer tool performs a full replication onto the target Polaris instance. This includes overwriting entities with the same name on the source and target, as well as removing entities from the target that do not exist on the source.

This change introduces a new option, `--strategy` to specify a level of modification the tool should perform on the target. There are three supported strategies:
* `CREATE_ONLY`: only create entities on the target instance, no overwrites, no removes
* `CREATE_AND_OVERWRITE`: create new entities on the target instance, overwrite entities with the same name across the source and target, no removes
* `REPLICATE`: full replication, including reflecting removals from the source instance onto the target

The lowest level of modification, `CREATE_ONLY` is the default. 

This PR also adds a flag `--delta-only` to make modification aware synchronization opt-in. If this flag is not specified,
* entities across the source and target will not be compared, just overwritten
* namespace metadata will not be compared across the source and target
* tables will not be loaded with etag header


travis-bowen (approved):
These changes make it a lot easier to configure with basic scenarios!

polaris-synchronizer/api/src/main/java/org/apache/polaris/tools/sync/polaris/PolarisSynchronizer.java
```diff
@@ -61,20 +61,24 @@ public class PolarisSynchronizer {
 
   private final boolean haltOnFailure;
 
+  private final boolean deltaOnly;
```
eric-maynard, line 64: Can we use a name other than `delta`? It sort of sounds like it only copies delta-related entities.

polaris-synchronizer/cli/src/main/java/org/apache/polaris/tools/sync/polaris/SyncPolarisCommand.java
```diff
@@ -120,11 +120,35 @@ public class SyncPolarisCommand implements Callable<Integer> {
...
 
+  @CommandLine.Option(
+          names = {"--delta-only"},
+          description = "Only synchronize the diff between the source and target Polaris."
+  )
+  private boolean deltaOnly;
+
+  @CommandLine.Option(
+          names = {"--strategy"},
+          defaultValue = "CREATE_ONLY",
```
eric-maynard, line 131: Is it possible to use the enum values here?

polaris-synchronizer/api/src/main/java/org/apache/polaris/tools/sync/polaris/PolarisSynchronizer.java
mansehajsingh: Renamed to `diffOnly`

polaris-synchronizer/cli/src/main/java/org/apache/polaris/tools/sync/polaris/SyncPolarisCommand.java
mansehajsingh: Unfortunately, we can't use the enum value. I could declare a separate set of constants for them that are assigned strings with the same value, if you think that's worth adding for maintainability? Something like
```java
enum Strategy {
    CREATE_ONLY,
    ...

    class Names {
         public static final String CREATE_ONLY = "CREATE_ONLY";
         ...
    }

}
```

mansehajsingh:
I wanna hold on this PR for a second. One thing I've thought of is that on overwrites we drop and recreate entities. I don't want to give the impression that we issue some kind of update request only.

polaris-synchronizer/cli/src/main/java/org/apache/polaris/tools/sync/polaris/SyncPolarisCommand.java
eric-maynard: Why doesn't e.g. `CREATE_ONLY.name()` work?

mansehajsingh: We can only use a String constant in an annotation.

eric-maynard: Oh, makes sense. Bummer

eric-maynard (approved):
LGTM, nice!

PR #20, opened 2025-04-29
Configurable StorageConfigInfo for Polaris benchmark

eric-maynard:
When trying to run the benchmark against a catalog that stores data in S3, I found that the StorageType is currently hardcoded to `FILE`.

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -232,6 +232,7 @@ The data set size can be adjusted as well. Each namespace is associated with an 
 
 The diagram below shows sample catalog, namespace and table definition given the following properties:
 
+-   storage-config-info: `{"storageType": "FILE"}`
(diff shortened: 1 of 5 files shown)


singhpk234 (commented):
LGTM ! Thank @eric-maynard 

lets update README as well

dimas-b (commented):
LGTM :+1:

@pingtimeout : could you have a look too?

pingtimeout (approved):
I am not a huge fan of injecting JSON.  But the alternative would be to duplicate the structure of Storage Config for each cloud vendor, and that would be unnecessarily complex.  So I think this implementation is the best way to address the requirement.  LGTM.

pingtimeout (approved):

dimas-b (approved):

PR #22, opened 2025-05-01
Disable`If-None-Match` when fetching tables in benchmark scenarios

eric-maynard:
I found gatling sometimes sending an etag to Polaris during testing, and this explicitly removes it and additionally disables caching for the HTTP protocol.

```
>>>>>>>>>>>>>>>>>>>>>>>>>>
Request:
Fetch Table: KO status.find.is(200), but actually found 304
=========================
Session:
Session(Reader-0-6,9,HashMap(gatling.http.cache.baseUrl -> http://localhost:8181, location -> s3://datalake-storage-team/emaynard/warehouse//C_0/NS_0/NS_2/NS_5/NS_12/NS_26/NS_53/NS_107/NS_216/T_358, catalogName -> C_0, multipartNamespace -> NS_0NS_2NS_5NS_12NS_26NS_53NS_107NS_216, tableUuid -> a2d9e974-1011-4a27-b2a7-27ea1c200d6b, be15ac8d-c58f-4b3f-9284-52615b346c84 -> 126, initialProperties -> HashMap(InitialAttribute_6 -> 6, InitialAttribute_1 -> 1, InitialAttribute_9 -> 9, InitialAttribute_7 -> 7, InitialAttribute_8 -> 8, InitialAttribute_0 -> 0, InitialAttribute_4 -> 4, InitialAttribute_3 -> 3, InitialAttribute_5 -> 5, InitialAttribute_2 -> 2), timestamp.be15ac8d-c58f-4b3f-9284-52615b346c84 -> 1746078977386, accessToken -> ..., gatling.http.cache.dns -> io.gatling.http.resolver.ShufflingNameResolver@5614ae36, gatling.http.cache.contentCache -> io.gatling.core.util.cache.Cache@2c670568, tableName -> T_358, gatling.http.ssl.sslContexts -> io.gatling.http.util.SslContexts@d180b99),KO,...
=========================
HTTP request:
GET http://localhost:8181/api/catalog/v1/C_0/namespaces/NS_0%1FNS_2%1FNS_5%1FNS_12%1FNS_26%1FNS_53%1FNS_107%1FNS_216/tables/T_358
headers:
        accept: application/json
        content-type: application/json
        Authorization: Bearer ...
        host: localhost:8181
        if-none-match: W/"4c5e0c65ecbacdeb95042b8d3a677b35060748a18a48b0cb333d5e5ae9b3aa50"
=========================
HTTP response:
version:
        HTTP/1.1
status:
        304 Not Modified
headers:
        content-encoding: identity
        content-length: 89
        Content-Type: application/json

<<<<<<<<<<<<<<<<<<<<<<<<<

```

### DIFF
--- a/benchmarks/src/gatling/scala/org/apache/polaris/benchmarks/actions/TableActions.scala
+++ b/benchmarks/src/gatling/scala/org/apache/polaris/benchmarks/actions/TableActions.scala
@@ -184,7 +184,7 @@ case class TableActions(
     http("Fetch Table")
       .get("/api/catalog/v1/#{catalogName}/namespaces/#{multipartNamespace}/tables/#{tableName}")
       .header("Authorization", "Bearer #{accessToken}")
-      .header("If-None-Match", null)
+      .header("If-None-Match", "")
       .check(status.is(200))
       .check(jsonPath("$.metadata.table-uuid").saveAs("tableUuid"))
       .check(jsonPath("$.metadata.location").is("#{location}"))
--- a/benchmarks/src/gatling/scala/org/apache/polaris/benchmarks/actions/TableActions.scala
+++ b/benchmarks/src/gatling/scala/org/apache/polaris/benchmarks/actions/TableActions.scala
@@ -184,6 +184,7 @@ case class TableActions(
     http("Fetch Table")
       .get("/api/catalog/v1/#{catalogName}/namespaces/#{multipartNamespace}/tables/#{tableName}")
       .header("Authorization", "Bearer #{accessToken}")
+      .header("If-None-Match", null)
       .check(status.is(200))
       .check(jsonPath("$.metadata.table-uuid").saveAs("tableUuid"))
       .check(jsonPath("$.metadata.location").is("#{location}"))
--- a/benchmarks/src/gatling/scala/org/apache/polaris/benchmarks/actions/TableActions.scala
+++ b/benchmarks/src/gatling/scala/org/apache/polaris/benchmarks/actions/TableActions.scala
@@ -184,7 +184,6 @@ case class TableActions(
     http("Fetch Table")
       .get("/api/catalog/v1/#{catalogName}/namespaces/#{multipartNamespace}/tables/#{tableName}")
       .header("Authorization", "Bearer #{accessToken}")
-      .header("If-None-Match", "")
       .check(status.is(200))
       .check(jsonPath("$.metadata.table-uuid").saveAs("tableUuid"))
       .check(jsonPath("$.metadata.location").is("#{location}"))
--- a/benchmarks/src/gatling/scala/org/apache/polaris/benchmarks/simulations/CreateTreeDataset.scala
+++ b/benchmarks/src/gatling/scala/org/apache/polaris/benchmarks/simulations/CreateTreeDataset.scala
@@ -149,6 +149,7 @@ class CreateTreeDataset extends Simulation {
     .baseUrl(cp.baseUrl)
     .acceptHeader("application/json")
     .contentTypeHeader("application/json")
+    .disableCaching
 
   // Get the configured throughput for tables and views
   private val tableThroughput = wp.createTreeDataset.tableThroughput
--- a/benchmarks/src/gatling/scala/org/apache/polaris/benchmarks/simulations/ReadTreeDataset.scala
+++ b/benchmarks/src/gatling/scala/org/apache/polaris/benchmarks/simulations/ReadTreeDataset.scala
@@ -152,6 +152,7 @@ class ReadTreeDataset extends Simulation {
     .baseUrl(cp.baseUrl)
     .acceptHeader("application/json")
     .contentTypeHeader("application/json")
+    .disableCaching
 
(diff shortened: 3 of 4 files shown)


benchmarks/src/gatling/scala/org/apache/polaris/benchmarks/actions/TableActions.scala
```diff
@@ -184,6 +184,7 @@ case class TableActions(
     http("Fetch Table")
       .get("/api/catalog/v1/#{catalogName}/namespaces/#{multipartNamespace}/tables/#{tableName}")
       .header("Authorization", "Bearer #{accessToken}")
+      .header("If-None-Match", "")
```
singhpk234, line 187: is the ETAG purely random or its in the format Polaris generates ? 
looks very similar to https://github.com/apache/polaris/blob/main/service/common/src/main/java/org/apache/polaris/service/http/IcebergHttpUtil.java#L45

eric-maynard: This is the same format that Polaris generates pretty much. From the etag example in the description, I think it's from Polaris. Unfortunately setting it as `""` did not clear it.

eric-maynard: https://groups.google.com/g/gatling/c/oYlzjJTAg1U

eric-maynard: Yeah I think that's exactly right @dimas-b

pingtimeout:
This is quite interesting, thanks for raising this @eric-maynard !

I think it can be solved in better ways.  The current fix consists in overwriting the header that Gatling puts, which feels not ideal.  What you could do is:

* Either replace `.check(status.is(200))` by `.check(status.in(200, 304))` call to make it clear that the HTTP 304 code is perfectly valid.  That better conveys the intent that Polaris may return that code.  This will be pulled by all simulations that fetch tables.
* Or update the `httpProtocol` definition in `ReadTreeDataset` and `ReadUpdateTreeeDataset` and add `.disableCaching` so that Gatling does not include the `If-None-Match` header anymore.  That effectively disables client caching and will force Polaris to return the full entity, leading to slightly more stress on the server.  And this would apply to all read queries, like namespaces and views, without having to overwrite the header everywhere.

I think `.disableCaching` is the way to go.  Wdyt?

eric-maynard:
> I think .disableCaching is the way to go. Wdyt?

The PR has this change already, wdym?

pingtimeout:
I found it weird that you linked the Gatling mailing list with the solution `.disableCaching` and suspected there was just one commit not pushed :-).  No worries.

Note that I think the header override is misleading you.  Like, you are seeing it in access logs, so it makes it look like the override is necessary, but the override is what creates the header itself.

If you run a workload with `.disableCaching` and just that, you will see that even though the server returns an Etag, Gatling does not send it in subsequent requests.  There is no need to override it.  See the output below with `TRACE` logging to see all queries.  Even though the server returns the Etag twice, the `If-None-Match` header is not sent in the second request for table `T_0`.

```
16:57:49.846 [TRACE] i.g.h.e.r.DefaultStatsProcessor - 
>>>>>>>>>>>>>>>>>>>>>>>>>>
Request:
Fetch Table: OK 
...
HTTP request:
GET http://localhost:8181/api/catalog/v1/C_0/namespaces/NS_0%1FNS_1%1FNS_3%1FNS_7%1FNS_15/tables/T_0
headers:
        accept: application/json
        content-type: application/json
        Authorization: Bearer ...
        host: localhost:8181
=========================
HTTP response:
version:
        HTTP/1.1
status:
        200 OK
headers:
        Content-Type: application/json;charset=UTF-8
        content-length: 1618
        ETag: W/"1aa1ed45b04ff32ecc20df4e9d7953a7cd71146c111fa6c8d8e3a9c147524f11"

[...]

16:57:50.455 [TRACE] i.g.h.e.r.DefaultStatsProcessor - 
>>>>>>>>>>>>>>>>>>>>>>>>>>
Request:
Fetch Table: OK 
...
HTTP request:
GET http://localhost:8181/api/catalog/v1/C_0/namespaces/NS_0%1FNS_1%1FNS_3%1FNS_7%1FNS_15/tables/T_0
headers:
        accept: application/json
        content-type: application/json
        Authorization: Bearer ...
        host: localhost:8181
=========================
HTTP response:
version:
        HTTP/1.1
status:
        200 OK
headers:
        Content-Type: application/json;charset=UTF-8
        content-length: 1618
        ETag: W/"1aa1ed45b04ff32ecc20df4e9d7953a7cd71146c111fa6c8d8e3a9c147524f11"
```

So I think the header override is unnecessary.

eric-maynard:
Fair enough @pingtimeout I will remove the explicit empty If-None-Match header! It's best that we rely on `disableCaching` and have scenarios add custom logic around etag when it's required. Thanks for taking a look

pingtimeout (approved):

PR #24, opened 2025-06-06
Add commits creator simulation

pingtimeout:
This new simulation adds a new workload that continuously updates the properties of every table and every view.  There is no bound to the maximum number of properties that can be created this way.  Each new property results in a new snapshot to be created for that particular entity.  This can quickly amount to hundreds of commits entity.

Using the `max-tables` and `max-views` properties, it is possible to concentrate commits creation to only a handful of entities, resulting in even bigger metadata sizes.

This simulation can serve as the basis for table metadata management workloads.

### DIFF
--- a/benchmarks/src/gatling/scala/org/apache/polaris/benchmarks/simulations/CreateCommits.scala
+++ b/benchmarks/src/gatling/scala/org/apache/polaris/benchmarks/simulations/CreateCommits.scala
@@ -95,7 +95,8 @@ class CreateCommits extends Simulation {
   val tableUpdateScenario: ScenarioBuilder =
     scenario("Create table commits by updating properties")
       .exec(authActions.restoreAccessTokenInSession)
-      .exec(feed(tableActions.propertyUpdateFeeder()).exec(tableActions.updateTable))
+      .feed(tableActions.propertyUpdateFeeder())
+      .exec(tableActions.updateTable)
 
   // --------------------------------------------------------------------------------
   // Read and write workloads:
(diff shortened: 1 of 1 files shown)


benchmarks/src/gatling/scala/org/apache/polaris/benchmarks/simulations/CreateCommits.scala
```diff
@@ -0,0 +1,138 @@
...
+  // --------------------------------------------------------------------------------
+  // Read and write workloads:
+  // * Create table commits by updating table properties
+  // * Read namespaces, tables and views
+  // --------------------------------------------------------------------------------
+  val tableUpdateScenario: ScenarioBuilder =
+    scenario("Create table commits by updating properties")
+      .exec(authActions.restoreAccessTokenInSession)
+      .feed(tableActions.propertyUpdateFeeder())
+      .exec(tableActions.updateTable)
```
dimas-b, line 99: I guess this makes a single table commit. It would be nice to have a multi-table commit simulation too (eventually) :)

dimas-b (approved):
Nice addition to the suite of workloads. Thanks, @pingtimeout !

benchmarks/src/gatling/scala/org/apache/polaris/benchmarks/simulations/CreateCommits.scala
dimas-b: Updating properties in multiple tables at the same time (the `/v1/{prefix}/transactions/commit` endpoint) is realistic enough for now :) That should exercise the code path that deals with atomically changing multiple entities in Persistence (same Persistence operation as for table data updates).

PR #27, opened 2025-09-22
Add missing incubation disclaimer

pingtimeout:
Note for reviewers: the ASF guidelines suggests that disclaimers should be set on a per-project basis, which I understand as a per-tool basis in the context of the `polaris-tools` repository.  I.e. instead of just adding a `DISCLAIMER` at the root of the repository, but not in each of the tools directory, I added one in each of the tools directory.  Let me know if it should be done differently.


jbonofre (commented):
It can go in an unique file in the root folder.

RussellSpitzer (approved):

jbonofre (approved):

PR #29, opened 2025-09-22
Add missing RAT task in benchmarks

pingtimeout:
This commit also includes a Spotless configuration so that missing headers are automatically added, along with adding a missing license header


dimas-b (approved):

singhpk234:
Thanks @pingtimeout !
Looks like we might have to run spotless :) on this ?

PR #32, opened 2025-10-03
Add polaris catalog migrator tests

ajantha-bhat:


iceberg-catalog-migrator/cli/src/test/java/org/apache/polaris/iceberg/catalog/migrator/cli/AbstractCLIMigrationTest.java
```diff
@@ -106,11 +109,6 @@ private static void initializeCatalog(
     }
   }
 
-  @AfterAll
-  protected static void tearDown() throws Exception {
```
ajantha-bhat, line 110: Child class need to shutdown the container after drop table now due to use of docker. So, this function cannot be at parent class. Moved to child classes.

iceberg-catalog-migrator/api-test/src/main/java/org/apache/polaris/iceberg/catalog/migrator/api/test/AbstractTest.java
```diff
@@ -55,8 +55,8 @@ public abstract class AbstractTest {
   protected static final Namespace NS_A_B_C_D = Namespace.of("a", "b", "c", "d");
   protected static final Namespace NS_A_B_C_D_E = Namespace.of("a", "b", "c", "d", "e");
 
-  private static String sourceCatalogWarehouse;
-  private static String targetCatalogWarehouse;
+  protected static String sourceCatalogWarehouse;
```
ajantha-bhat, line 58: Using them in child class

iceberg-catalog-migrator/gradle/libs.versions.toml
```diff
@@ -75,6 +76,7 @@ logback-classic = { module = "ch.qos.logback:logback-classic", version.ref = "lo
 logcaptor = { module = "io.github.hakky54:logcaptor", version.ref = "logcaptor" }
 picocli = { module = "info.picocli:picocli", version.ref = "picocli" }
 slf4j = { module = "org.slf4j:log4j-over-slf4j", version.ref = "slf4j" }
+testcontainers-bom = { module = "org.testcontainers:testcontainers-bom", version.ref = "testcontainers" }
```
dimas-b, line 79: Is it used?

dimas-b (approved):
LGTM overall 👍 Thanks for adding these tests, @ajantha-bhat !

ajantha-bhat:
Locally the individual tests have passed. Need to debug now on why it has failed on PR builder.

iceberg-catalog-migrator/gradle/libs.versions.toml
ajantha-bhat: testcontainer dependency needed as 

`PolarisContainer extends GenericContainer<PolarisContainer> {`

So, added a dependency in https://github.com/apache/polaris-tools/pull/32/files#diff-63fbf4e3b782dbcabcab9699b864aea9b96668ec6215bc8951f8cb6ba2223e66R131

Did you mean why depend on bom instead of a specific dependency version?

iceberg-catalog-migrator/cli/src/test/java/org/apache/polaris/iceberg/catalog/migrator/cli/AbstractCLIMigrationTest.java
```diff
@@ -116,12 +114,6 @@ protected void beforeEach() {
 
   @AfterEach
   protected void afterEach() {
-    // manually refreshing catalog due to missing refresh in Nessie catalog
-    // https://github.com/apache/iceberg/pull/6789
-    // create table will call refresh internally.
-    sourceCatalog.createTable(TableIdentifier.of(BAR, "tblx"), schema).refresh();
```
ajantha-bhat, line 124: Since for polaris, source and target catalog uses same warehouse (Needed for migration). This was failing in CI when target catalog was trying to create table as the same path table exists. 

Now. I removed this as the above mentioned Nessie issue (in the comment) is fixed and we don't need this workaround.

iceberg-catalog-migrator/gradle/libs.versions.toml
ajantha-bhat: Aah. Done. Removed.

PR #72, opened 2025-11-23
[MCP] feat: add retry for HTTP requests

MonkeyCanCode:
Currently there is no retry if there is a connection error between mcp server and Polaris server. This PR adds support for configurable retry.

### DIFF
--- a/mcp-server/polaris_mcp/server.py
+++ b/mcp-server/polaris_mcp/server.py
@@ -90,8 +90,6 @@ def create_server() -> FastMCP:
     """Construct a FastMCP server with Polaris tools."""
     base_url = _resolve_base_url()
     timeout = _resolve_http_timeout()
-    http = urllib3.PoolManager()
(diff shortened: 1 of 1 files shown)


mcp-server/polaris_mcp/server.py
```diff
@@ -59,13 +59,30 @@
     "required": ["isError"],
     "additionalProperties": True,
 }
+DEFAULT_HTTP_RETRIES_TOTAL = 3
+DEFAULT_HTTP_RETRIES_BACKOFF_FACTOR = 0.5
+HTTP_RETRIES_STATUS_FORCELIST = [401, 409, 429, 500, 502, 503, 504]
```
flyrain: I think the write API retry should be more cautious. @singhpk234 has a PR in Iceberg to not retry on 50x.

flyrain: Not a blocker: we could retry for read endpoint with 5xx. We don't have to change that now. And we might not need to handle retry once we integrate with pyIceberg.

PR #73, opened 2025-11-23
[MCP] feat: add support for configurable HTTP timeout

MonkeyCanCode:
This PR adds support for configurable HTTP timeout. The current code is using `20.0` for auth for both connect/read and `30.0` for all of the rest. This PR default everything to `30.0` if not specify and adds support to control these via `POLARIS_HTTP_TIMEOUT_SECONDS`, `POLARIS_HTTP_CONNECT_TIMEOUT_SECONDS`, and `POLARIS_HTTP_READ_TIMEOUT_SECONDS`

### DIFF
--- a/mcp-server/polaris_mcp/server.py
+++ b/mcp-server/polaris_mcp/server.py
@@ -83,7 +83,6 @@ LOGGING_CONFIG = {
 logger = logging.getLogger(__name__)
 
 
-
 def create_server() -> FastMCP:
     """Construct a FastMCP server with Polaris tools."""
     base_url = _resolve_base_url()


adam-christian-software (approved):
🚀

flyrain (approved):
LGTM! Thanks a lot for working on it @MonkeyCanCode ! 
A side note: looks like we are re-implement a lot of stuff from pyIceberg. Should we integration pyIceberg as much as possible? For example, It seems a good idea to based on the pyIceberg AuthManager https://py.iceberg.apache.org/reference/pyiceberg/catalog/rest/auth/.

flyrain:
> Very good point. Yeah, let me look into pyiceberg code base a bit later this weekend then see what I can do there. I only scan through pyiceberg code base couple months back for couple modules for a different purpose. Maybe we can merge for now then refactor as the follow up PR?

Feel free to merge it. Was brainstorming with you, not a blocker.

PR #79, opened 2025-11-27
[Synchronizer] chore: Add github workflow for synchronizer

MonkeyCanCode:
Currently synchronizer doesn't have have github workflow and there is a missing file for gradle and outdated content for gradlew. This PR addresses these issues.


flyrain (approved):
Thanks for the fix!

PR #85, opened 2025-11-30
Introducing Apache Polaris Console

jbonofre:
This PR is proposing the Apache Polaris Console, an UI to browse and manage Polaris server and entities.

A huge thanks to the Console contributors: @sohanhonavar and @binarycat0

### DIFF
--- a/console/helm/Chart.yaml
+++ b/console/helm/Chart.yaml
@@ -1,3 +1,22 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#  http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
 apiVersion: v2
 name: polaris-console
 type: application
--- a/console/helm/templates/_helpers.tpl
+++ b/console/helm/templates/_helpers.tpl
@@ -1,4 +1,25 @@
 {{/*
+
+  Licensed to the Apache Software Foundation (ASF) under one
+  or more contributor license agreements.  See the NOTICE file
+  distributed with this work for additional information
+  regarding copyright ownership.  The ASF licenses this file
+  to you under the Apache License, Version 2.0 (the
+  "License"); you may not use this file except in compliance
+  with the License.  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing,
+  software distributed under the License is distributed on an
+  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  KIND, either express or implied.  See the License for the
+  specific language governing permissions and limitations
+  under the License.
+
+*/}}
+
+{{/*
 Expand the name of the chart.
 */}}
 {{- define "polaris-console.name" -}}
--- a/console/helm/templates/deployment.yaml
+++ b/console/helm/templates/deployment.yaml
@@ -1,3 +1,22 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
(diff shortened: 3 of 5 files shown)


console/src/api/auth.ts
```diff
@@ -0,0 +1,131 @@
...
+import axios from "axios"
+import { apiClient } from "./client"
+import { navigate } from "@/lib/navigation"
+import { REALM_HEADER_NAME } from "@/lib/constants"
+import type { OAuthTokenResponse } from "@/types/api"
+
+// Always use relative URL to go through the proxy (dev server or production server)
+// This avoids CORS issues by proxying requests through the server
+// The server.ts proxy handles /api routes in production, and Vite handles them in development
+const TOKEN_URL = "/api/catalog/v1/oauth/tokens"
```
flyrain: Can we have a configurable token url for 3-rd party IDPs or different remote Polaris instance? Not a blocker. I'm fine with filing an issue and resolving it later.

console/src/api/catalog/namespaces.ts
```diff
@@ -0,0 +1,137 @@
...
+  list: async (
+    prefix: string,
+    parent?: string[]
+  ): Promise<Namespace[]> => {
+    const params: Record<string, string> = {}
+    if (parent && parent.length > 0) {
+      params.parent = encodeNamespace(parent)
+    }
+
+    const response = await apiClient
```
MonkeyCanCode: NIT: should we consider for adding support for `pageSize` as if an user create high number of namespace, the UI loading will taking extremely long of time.

console/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+# Build stage
+FROM node:20-alpine AS builder
+
+WORKDIR /app
+
+# Copy package files
+COPY package*.json ./
+
+# Install dependencies
+RUN npm ci --only=production=false
```
MonkeyCanCode:
```
 > [builder 4/6] RUN npm ci --only=production=false:
0.192 npm warn invalid config only="production=false" set in command line options
```

doesn't appear to be an valid option. Maybe we should do `RUN npm ci --production=false` instead?

console/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+# specific language governing permissions and limitations
+# under the License.
+
+# Build stage
+FROM node:20-alpine AS builder
+
+WORKDIR /app
+
+# Copy package files
+COPY package*.json ./
```
MonkeyCanCode: From the checked in code, the `package-lock.json` seems to be missing which will fail the next command.

console/package.json
```diff
@@ -0,0 +1,58 @@
+{
+  "name": "polaris-console",
+  "private": true,
+  "version": "1.3.0-incubating",
+  "type": "module",
+  "scripts": {
```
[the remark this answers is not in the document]
adam-christian-software: Sounds good! I just want to ensure they are tracked. Thanks, @sohanhonavar & @jbonofre, y'all can close this comment when you track it on your side. :)

dimas-b (approved):
I tested the UI locally and it worked fine and looked great :) I do not have much to say about the code itself as I'm not familiar with the framework.

console/src/api/auth.ts
```diff
@@ -0,0 +1,131 @@
...
+
+    if (response.data.access_token) {
+      apiClient.setAccessToken(response.data.access_token)
+    }
+
+    return response.data
+  },
+
+  logout: (): void => {
+    localStorage.removeItem("polaris_access_token")
```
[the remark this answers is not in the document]
flyrain: Can we file an issue for this? Thanks!

MonkeyCanCode (approved):

MonkeyCanCode:
Local testing works now with latest commit and I am able to integrated it with Polaris server in helm as well. There are couple files missed license headers. Other than this, same as @dimas-b , I am not very familiar with the framework as well.

binarycat0:
Hi @georgelza! Great ideas 👍, IMO it's better to create a new GH Issue with proposal type or even start a new dev-mailing thread and continue discussion there 🙏

console/src/pages/Connections.tsx
```diff
@@ -0,0 +1,368 @@
...
+                    : typeof principalWithTimestamps.properties?.lastUpdateTimestamp === "number"
+                    ? principalWithTimestamps.properties.lastUpdateTimestamp
+                    : undefined)
+
+                return (
+                  <TableRow key={principal.name}>
+                    <TableCell className="font-medium">{principal.name}</TableCell>
+                    <TableCell>
+                      <Badge variant="secondary">
+                        {principal.properties?.serviceType || "apache-spark"}
```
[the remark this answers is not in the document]
flyrain: Thanks for the clarification. 
I still have questions to better understand the UX and conceptual model here. Do we really need a separate service type concept in the UI?
1. From a user’s perspective, this seems to overlap with the existing principal concept already exposed on the access control page.
2. How do we expect users to use “service type”? Given the only difference between “principal” and “service type” is having a property like `service-type=Spark`

I’m trying to understand whether this duplication is intentional and beneficial from a UX perspective, or if it risks confusing users.

flyrain (approved):
Thanks a lot for working on it. This a great start point. I'd suggest to merge it first and file related issues.

Issue #131, opened 2026-01-10
Lack of renovate bot to keep dependencies up to date


MonkeyCanCode:
Unlike https://github.com/apache/polaris, there is no renovate bot to keep project dependencies up to date in this repo. This can make many of the tools hard to be used/adopted due to various known CVEs from dependencies.

PR #132, opened 2026-01-11
Apply the same renovate configuration between polaris and polaris-tools

jbonofre:


.github/renovate.json5
```diff
@@ -1,42 +1,94 @@
...
+      groupName: "Quarkus Platform and Group",
+      matchManagers: ["gradle"],
+      matchPackageNames: [
+        "io.quarkus.platform:quarkus-bom",
+        "io.quarkus.platform:quarkus-amazon-services-bom",
+        "io.quarkus.platform:quarkus-google-cloud-services-bom",
+        "io.quarkus:io.quarkus.gradle.plugin",
+      ],
     },
+    {
```
MonkeyCanCode: This section can be remove as well as that was added to prevent bump version of `openapi-generator-cli` as we can't just blindly go to the latest one due to OpenAPI spec we are using but this is not used in this repo.

MonkeyCanCode:
@jbonofre thanks for the quick fix for https://github.com/apache/polaris-tools/issues/131. Left couple comments on un-used sections that we may want to consider remove as those are not being used in this repo.

.github/renovate.json5
```diff
@@ -1,42 +1,94 @@
...
+  // BE CAREFUL WITH THIS OPTION and make sure that the changes work.
+  baseBranches: [
+    "main",
+  ],
 
-  "labels": ["dependencies"],
+  pip_requirements: {
+    // fileMatch default: (^|/)([\\w-]*)requirements\\.(txt|pip)$
+    "fileMatch": ["(^|/)([\\w-]*)requirements.*\\.txt$"],
+  },
```
snazy: This can be removed (there's no `requirements*.txt`).

snazy:
... rather on getting renovate in

snazy:
Ah, bummer. Looks like I opened a similar PR #170.  What shall we do?

PR #133, opened 2026-01-12
Catalog Migrator: Remove AWS SDK dependencies for runtime

ajantha-bhat:
Runtime jar was 650 MB because of this. Now it is around 118 MB (still packs hadoop-aws). 

Initially wanted to have it has standalone jar, but we were only depending on AWS SDK. We need other GCP or AZURE dependencies if we want to be standalone jar for all systems. Thats will bloat up the runtime jar size. Hence, excluded the dependencies and user need to provide them in the classpath based on the storage type (similar to runtime jars of Iceberg)


ajantha-bhat:
ping @jbonofre

ajantha-bhat:
I will be merging this PR in some time today if not more comments and do an RC cut as the changes were reviewed and we need to do the release soon.

jbonofre (approved):

Issue #125, opened 2026-01-08
[Feature Request] [Console] Add Authorization Option via OIDC


binarycat0:
### Motivation
Enable secure authentication via an external OpenID Connect (OIDC) Identity Provider (e.g. Keycloak) for Console UI application.

Using `grant_type=password` in browser-based apps is discouraged, as it requires exposing user credentials and/or client secrets.  
**Authorization Code Flow with PKCE** is the recommended and secure approach for SPAs.

### Proposed Configuration
Add optional OIDC settings:

- `OIDC_ISSUER_URL`
- `OIDC_CLIENT_ID`
- `OIDC_REDIRECT_URI`

### Proposed Behavior
If OIDC is configured:

- Show a **“Login with OIDC”** button on the login page
- Redirect the user to the external IdP login page
- After successful authentication, redirect back to `OIDC_REDIRECT_URI`
- On the callback page:
  - Exchange the authorization code for tokens using **Authorization Code + PKCE**
  - Store the access token in memory/session
  - Use the token for authenticated API requests

### Security Notes
- No client secret stored in the UI
- User credentials are handled only by the IdP
- Aligned with OAuth 2.1 / OIDC SPA best practices

### Suggested Implementation
- JS library: `react-oidc-context`

### Alternative (Optional)
Implement a middleware / BFF layer:
- UI performs redirect-based login only
- Middleware handles token exchange and server-to-server authentication
- Prevents exposing tokens or secrets in the browser

cc: @jbonofre @sohanhonavar

binarycat0:
An alternative implementation of the OIDC integration for Console UI without "Workspaces". 
- https://github.com/apache/polaris-tools/pull/168

PR #138, opened 2026-01-23
feat: introduce workspaces

binarycat0:
## Motivation

The console previously relied on environment variables to configure a single Polaris context (server, realm, and authentication). This approach does not align with the idea of a management console that can be used across different realms and environments, and it limits flexibility when working with multiple Polaris setups or experimenting with authentication methods.

Introducing **workspaces** shifts the focus from static, environment-based configuration to a user-managed model. A workspace represents an isolated Polaris context (server + realm + auth), making it possible to use one console instance to interact with multiple Polaris environments in a consistent and scalable way.

This change is also a preparatory step for upcoming OIDC integration in the Console UI:  
[WIP] https://github.com/apache/polaris-tools/issues/125

## Features

- Support for multiple **workspaces**, each representing a distinct Polaris context
- Workspace configuration managed via the UI instead of environment variables
- Each workspace encapsulates:
  - Polaris server URL
  - Realm and realm header name
  - Authentication configuration
- Ability to authenticate into multiple workspaces and switch between them
- Multiple authentication model (currently client credentials; OIDC unblocked for future work)

## Benefits

- Use a single console instance to work with multiple Polaris environments
- Clean separation of realms and authentication contexts via workspaces
- No need to reconfigure or restart the console when switching environments
- Improved UX through explicit workspace selection and session persistence
- Establishes a clear foundation for OIDC and additional auth providers

## Additional Details

- Removed dependency on environment variables for realm and auth configuration
- Added workspace selection to the login flow
- Authentication state is scoped per workspace
- Design intentionally aligns workspaces with future extensibility (auth providers, policies, UX improvements)

cc: @jbonofre @sohanhonavar @adam-christian-software @dimas-b 

----

## Modified login form
image

## Ability to chose between the multiple workspaces and auth methods
image

## Dummy warning about WIP on the OIDC implementation
image

## Workspaces Management space
image

## Show active worskpace. Switching between workspaces
image


dimas-b (commented):
Very nice improvement, @binarycat0 ! I think this feature is worth adding to Polaris UI.

I cannot say much about actual code change, but I ran this UI in my local and it worked very well. Some UX notes below.

* In the "create workspace" there are auto-filled values, but I do not see an easy way to accept them... the `Create` button does not work unless I type in the same values manually.
* When under Home it's not clear what workspace I'm in. Current Realm is shown, though, but it's not 1:1 with workspace.
* `Workspaces` show all possible workspaces, but on the left-hand pane this page is parallel to `Catalogs`, while `Catalogs` is specific to the current workspace, I assume, right? This is kind of awkward. I'd proposed to remove the `Workspaces` page from the left-hand pane and only show the current workspace when the user is logged in.

dimas-b (dismissed):
Thanks for the update, @binarycat0 ! The workspaces UI LGTM now, but I hope someone more familiar with the framework can review too... 

@jbonofre ?

PR #150, opened 2026-01-30
Bump Gradle to 9.3.1 in all tools

snazy:


apprunner/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,27 +1,7 @@
-#
```
snazy: The license header would be removed during a Gradle bump by Renovate (via `gradle wrapper ...`).

PR #209, opened 2026-04-05
feat(Console): Add quick setup feature

vignesh-manel:
Adds an interactive wizard to the console that lets users bootstrap Polaris from scratch — creating principal roles, principals, catalogs, namespaces, and catalog roles in one guided flow using either preset examples or user-provided YAML.

## Features & Changes

 - Two setup modes — choose a ready-made preset (Local File, AWS S3, Azure ADLS, Google Cloud) or paste/upload a custom YAML config in the same format as `polaris setup apply`
 - Guided preset flow — parameterized inputs (catalog name, storage location, role ARN, etc.) with required-field validation before proceeding
 - YAML mode with live validation — 300ms debounce parse with inline feedback; validates structure, required fields (storage_type, default_base_location), and storage type values
 - Step preview — review all planned actions before applying; existing entities are automatically skipped (409 treated as success)
 - Live progress view — each step updates in real-time as it executes; failed steps are clearly flagged while remaining steps continue where possible
 - Credential display — generated principal credentials are shown with copy buttons at the end of the flow; not persisted anywhere
 - Error resilience — catalog sub-steps are skipped gracefully if catalog creation fails; a fatal API error (e.g. 403 on list) halts remaining steps cleanly with a summary

## Limitations / Known Issues

 - No partial resume — if the wizard is closed mid-execution, there is no way to resume; re-running from the start is safe (idempotent) but will skip already-created entities
 - YAML validation is structural only — the wizard validates shape and required fields but does not verify that referenced roles, catalogs, or principals actually exist in Polaris before running
 - Preset YAML view is read-only — the generated config from a preset cannot be edited in the wizard before applying; switch to YAML mode for full control
 - Credentials shown once — generated client secrets are displayed only in the final step and are not recoverable after closing the wizard
 - No teardown support — the wizard only creates entities; there is no corresponding flow to delete or clean up the resources it provisioned

https://github.com/user-attachments/assets/299cc61b-d326-42fb-8780-64253930875c

### DIFF
--- a/console/package.json
+++ b/console/package.json
@@ -30,6 +30,7 @@
     "clsx": "^2.1.1",
     "cmdk": "^1.1.1",
     "date-fns": "^4.1.0",
+    "js-yaml": "^4.1.1",
     "lucide-react": "^0.548.0",
     "react": "^19.2.0",
     "react-dom": "^19.2.0",
@@ -42,6 +43,7 @@
   },
   "devDependencies": {
     "@eslint/js": "^9.36.0",
+    "@types/js-yaml": "^4.0.9",
     "@types/node": "^24.9.2",
     "@types/react": "^19.1.16",
     "@types/react-dom": "^19.1.9",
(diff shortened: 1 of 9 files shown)


vignesh-manel:
@jbonofre Requesting for PR review whenever you have time. Thanks

Issue #197, opened 2026-03-22
FastMCP HTTP/SSE transport does not forward incoming request headers (e.g. Authorization) to the Polaris backend


creepysta:
When the MCP server is running in `http` or `sse` transport mode, incoming HTTP headers from the client request are **never extracted or forwarded** to the backend Polaris REST API. This makes it impossible to deploy the MCP server behind an authenticating reverse proxy (e.g. OAuth2 Proxy) and have the proxy-injected `Authorization` header passed through to Polaris.

### Use Case Scenario

```
Client → OAuth2 Proxy → Polaris MCP Server (http transport) → Polaris Server (external IDP)
```

1. The MCP server is deployed **without its own auth** (`POLARIS_CLIENT_ID` / `POLARIS_CLIENT_SECRET` are not set; the Polaris server has client credentials disabled).
2. An authenticating reverse proxy (e.g. [OAuth2 Proxy](https://oauth2-proxy.github.io/oauth2-proxy/)) sits in front of the MCP server.
3. The proxy handles the OAuth2/OIDC redirect flow with the external Identity Provider and injects an `Authorization: Bearer <token>` header into every request forwarded to the MCP server.
4. The MCP server should forward that `Authorization` header to the backend Polaris server, which validates the token against the same external IDP.

**This flow currently does not work** because the MCP server never reads headers from the incoming HTTP request.

### Current Behavior

The auth resolution in `server.py` (`_resolve_authorization_provider`) only considers:

1. **Static tokens** from env vars (`POLARIS_API_TOKEN`, `POLARIS_BEARER_TOKEN`, `POLARIS_TOKEN`)
2. **OAuth2 Client Credentials** from env vars (`POLARIS_CLIENT_ID` + `POLARIS_CLIENT_SECRET`)
3. **Realm-specific credentials** from `POLARIS_REALM_*` env vars

If none of these are set, the `AuthorizationProvider` resolves to `none()` — no `Authorization` header is added to outbound requests, and the incoming HTTP-layer `Authorization` header is silently dropped.

The `headers` parameter on each tool function (e.g. `polaris_iceberg_table`) only accepts headers passed **as MCP tool arguments by the LLM client**, not headers from the HTTP transport layer. 

--- 

We should be able to use the `get_http_headers` from fast mcp in `RestTool::call()`.

creepysta:
I'm happy to take it up. Please let me know.

MonkeyCanCode:
[...]

PR #215, opened 2026-04-17
fix: forward auth header from http request

creepysta:
https://github.com/apache/polaris-tools/issues/197

Currently the mcp server can only be configured with auth with static tokens, or client credentials.

But it doesn't forward auth headers from clients connecting to it. This becomes important when both the client and catalog server is configured to talk to an external IDP (`polaris.authentication.type=external`) 

This PR allows the MCP server to simply forward the `Authorization` header (if its present in the http request) from the client to the catalog server.

### DIFF
--- a/mcp-server/polaris_mcp/rest.py
+++ b/mcp-server/polaris_mcp/rest.py
@@ -27,6 +27,7 @@ from typing import Any, Dict, List, Optional, Tuple
 from urllib.parse import urlencode, urljoin, urlsplit, urlunsplit, quote
 
 import urllib3
+from fastmcp.server.dependencies import get_http_headers
 
 from polaris_mcp.authorization import AuthorizationProvider, none
 from polaris_mcp.base import JSONDict, ToolExecutionResult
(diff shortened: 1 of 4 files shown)


flyrain (commented):
Thanks @creepysta ! LGTM. Can we add a test for this to avoid any regression in the future?

creepysta:
@flyrain , I have added the tests. 
One thing to note, looks like the dependency version resolved in my local was `3.1.0` although the pinned dep is `>=2.13.0`. 
The following is the commit which had introduced stripping the `Authorization` header, which was introduced in fastmcp 3.0.2 (earliest) 
https://github.com/PrefectHQ/fastmcp/commit/0d4580fef32a8666b93dabbef86e05d763307ba4

Therefore have also bumped the version of fastmcp to `>=3.1.0`. Tests are passing.

flyrain (approved):

creepysta:
Thanks for the review @flyrain! All checks seem to have passed. I do not have access to merge the PR, please help merging it.

flyrain:
Thanks @creepysta for the contribution!

PR #225, opened 2026-05-12
Add overwrite support for migrator

MonkeyCanCode:
Currently the migrator supports both `register` and `migrate` which are great. However, for use cases where people would register the tables and let them sit idle for couple of days then data ops kicked in on the source catalog which removed the previous metadata pointers, this can caused the target tables to be not read-able as it would fails when looking up metadata file from metadata pointer.

To overcome this issue, I added `--overwrite` option which would re-register a table if and only if the metadata pointer changed.

### DIFF
--- a/iceberg-catalog-migrator/.gitignore
+++ b/iceberg-catalog-migrator/.gitignore
@@ -17,6 +17,69 @@
 # under the License.
 #
 
+# Ignore Gradle GUI config
+gradle-app.setting
+
 # Ignore Gradle wrapper jar file
 gradle/wrapper/gradle-wrapper.jar
 gradle/wrapper/gradle-wrapper-*.sha256
+
+# Avoid ignore Gradle wrappper properties
+!gradle-wrapper.properties
+
+# Cache of project
+.gradletasknamecache
+
+# Eclipse Gradle plugin generated files
+# Eclipse Core
+.project
+# JDT-specific (Eclipse Java Development Tools)
+.classpath
+.env
+
+# IntelliJ
+.idea/
+*.iml
+*.ipr
+*.iws
+
+# Gradle
+/.gradle
+/build-logic/.gradle
+/build-logic/.kotlin
+**/build/
+!src/**/build/
+
+# jenv
+.java-version
+
+# binary files
+*.class
+*.jar
+*.zip
+*.tar.gz
+*.tgz
+
+# virtual machine crash logs, see http://www.java.com/en/download/help/error_hotspot.xml
+hs_err_pid*
+
+# macOS
+*.DS_Store
+
+# vscode
+.vscode/
+
+# Python Virtual Env
+.venv
+venv
+
+# AI / Agentic Development Tools
+.agents/
+.aider/
+.aider.chat.history.md
+.aider.input.history
+.codex/
+.claude/
+.cursor/
+.opencode/
+skills-lock.json
--- a/iceberg-catalog-migrator/api/src/main/java/org/apache/polaris/iceberg/catalog/migrator/api/CatalogMigrator.java
+++ b/iceberg-catalog-migrator/api/src/main/java/org/apache/polaris/iceberg/catalog/migrator/api/CatalogMigrator.java
@@ -53,6 +53,15 @@ public abstract class CatalogMigrator {
   /** Delete the table entries from the source catalog after successful registration. */
   public abstract boolean deleteEntriesFromSourceCatalog();
 
+  /**
+   * Re-register the table in the target catalog if its metadata location has diverged from source.
+   * *
+   */
+  @Value.Default
+  public boolean overwrite() {
+    return false;
+  }
+
   /** Enable the stacktrace in logs in case of failures. */
   @Value.Default
   public boolean enableStacktrace() {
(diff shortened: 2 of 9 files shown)


iceberg-catalog-migrator/api/src/main/java/org/apache/polaris/iceberg/catalog/migrator/api/CatalogMigrator.java
```diff
@@ -211,11 +220,42 @@ protected void getAllNamespacesFromSourceCatalog(Namespace namespace, Set<Namesp
...
+                "Table {} is already in sync with the target catalog. Skipping re-registration.",
+                tableIdentifier);
+            return true;
+          }
+
+          LOG.info(
+              "Table {} metadata location has changed. Dropping and re-registering it in the target catalog as overwrite is enabled.",
+              tableIdentifier);
+          targetCatalog().dropTable(tableIdentifier, false);
+          targetCatalog().registerTable(tableIdentifier, sourceMetadataLocation);
```
ajantha-bhat, line 249: We don't have to handle like this. 

Iceberg provides an API with the force register option already. 

Refer this: 
https://github.com/apache/iceberg/blob/3d682a3b65838aba29f16141b45633a6cbe51178/api/src/main/java/org/apache/iceberg/catalog/Catalog.java#L362

iceberg-catalog-migrator/build.gradle.kts
```diff
@@ -76,6 +76,9 @@ tasks.named<RatTask>("rat").configure {
 
   // Rat can't scan binary images
   excludes.add("**/*.png")
+
+  // Misc build artifacts
+  excludes.add(".java-version")
```
ajantha-bhat, line 81: unrelated changes to this PR?

iceberg-catalog-migrator/.gitignore
```diff
@@ -17,6 +17,69 @@
...
 #
 
+# Ignore Gradle GUI config
+gradle-app.setting
+
 # Ignore Gradle wrapper jar file
 gradle/wrapper/gradle-wrapper.jar
 gradle/wrapper/gradle-wrapper-*.sha256
+
+# Avoid ignore Gradle wrappper properties
```
ajantha-bhat, line 27: Unrelated changes to this PR?

iceberg-catalog-migrator/docs/examples.md
```diff
@@ -120,3 +120,14 @@ java -jar iceberg-catalog-migrator-cli-0.0.1.jar migrate \
 
 ## Tips
 1. Before migrating tables to Polaris, make sure the catalog is configured to the `base-location` same as source catalog `warehouse` location during catalog creation.
+2. If you have already registered tables and the source catalog metadata has changed (e.g., after table maintenance or new commits), you can use the `--overwrite`
```
ajantha-bhat, line 123: This implies that user is using the same table from multiple catalogs. Which is not recommended. 

See here about the warning: https://github.com/apache/polaris-tools/tree/main/iceberg-catalog-migrator#readme

ajantha-bhat (commented):
Thanks for working on this. 
Implementation can be simple (provided more info in comments). 

⚠️  I think the intention here is that user is using same table from multiple catalogs. 
Which is strongly discouraged as mentioned in the docs here. 

https://github.com/apache/polaris-tools/tree/main/iceberg-catalog-migrator#readme

So, I am not strongly convinced about the use case here.

iceberg-catalog-migrator/build.gradle.kts
MonkeyCanCode: This is to match to the jenv that we are using on the Polaris. I can move them to a different PR if preferred.

MonkeyCanCode:
> Thanks for working on this. 
> Implementation can be simple (provided more info in comments). 
> 
> ⚠️  I think the intention here is that user is using same table from multiple catalogs. 
> Which is strongly discouraged as mentioned in the docs here. 
> 
> https://github.com/apache/polaris-tools/tree/main/iceberg-catalog-migrator#readme
> 
> So, I am not strongly convinced about the use case here. 

I do understand why this is discouraged as there is a potential a write can happen on both end (or table maintenances but this will be more like setup issues IMO). However, for large iceberg lakehouse that have multiple writers and they are responsible for a set of tables, it is not very possible to just migrate it all at one time. In that case, we do need this type of functionality to re-register a table. WDYT?

ajantha-bhat:
> I do understand why this is discouraged as there is a potential a write can happen on both end (or table maintenances but this will be more like setup issues IMO). However, for large iceberg lakehouse that have multiple writers and they are responsible for a set of tables, it is not very possible to just migrate it all at one time. In that case, we do need this type of functionality to re-register a table. WDYT?

Can we migrate batch by batch for this scenario? single batch can be all tables in one namespace. 
This tool supports identifier regex. So, we can use that to migrate specific group of tables. So, no need to attempt re-register IMO. 

Also, I would like to add that, if only 'K' tables migrated out of 'N' and you need to migrate remaining. You can call same cli for all 'N' tables. It will skip migration for registered tables (with the error that table already exist). But other tables will still be migrated in that process. It won't pause migration of other tables if current table already exists.

MonkeyCanCode:
> > I do understand why this is discouraged as there is a potential a write can happen on both end (or table maintenances but this will be more like setup issues IMO). However, for large iceberg lakehouse that have multiple writers and they are responsible for a set of tables, it is not very possible to just migrate it all at one time. In that case, we do need this type of functionality to re-register a table. WDYT?
> 
> 
> 
> Can we migrate batch by batch for this scenario? single batch can be all tables in one namespace. 
> 
> This tool supports identifier regex. So, we can use that to migrate specific group of tables. So, no need to attempt re-register IMO. 
> 
> 
> 
> Also, I would like to add that, if only 'K' tables migrated out of 'N' and you need to migrate remaining. You can call same cli for all 'N' tables. It will skip migration for registered tables (with the error that table already exist). But other tables will still be migrated in that process. It won't pause migration of other tables if current table already exists. 

So this may not be 100% valid as people can be using namespace as tenant/workflow separator for a given writer. Assuming one app is writing to 1k tables and each of the table has their own namespace, the purposed solution may not be useable in this case.

ajantha-bhat:
> So this may not be 100% valid as people can be using namespace as tenant/workflow separator for a given writer. Assuming one app is writing to 1k tables and each of the table has their own namespace, the purposed solution may not be useable in this case.

I can understand that namespace by namespace migration is not possible in this case. 

What about retrying again for failed tables like mentioned below? 

> Also, I would like to add that, if only 'K' tables migrated out of 'N' and you need to migrate remaining. You can call same cli for all 'N' tables. It will skip migration for registered tables (with the error that table already exist). But other tables will still be migrated in that process. It won't pause migration of other tables if current table already exists.

MonkeyCanCode:
> > So this may not be 100% valid as people can be using namespace as tenant/workflow separator for a given writer. Assuming one app is writing to 1k tables and each of the table has their own namespace, the purposed solution may not be useable in this case.
> 
> I can understand that namespace by namespace migration is not possible in this case.
> 
> What about retrying again for failed tables like mentioned below?
> 
> > Also, I would like to add that, if only 'K' tables migrated out of 'N' and you need to migrate remaining. You can call same cli for all 'N' tables. It will skip migration for registered tables (with the error that table already exist). But other tables will still be migrated in that process. It won't pause migration of other tables if current table already exists.

Hello @ajantha-bhat , thanks for the info and sorry for the late response. For sure it is possible to just keep rerun the migrate to get all tables over. But this doesn't necessary cover the case where people who want to benchmark reads on two diff catalogs (point read-only operation to the new catalog). I do understand the concern with two tables registered on two catalogs can be dangerous as there can be split-brain problem, thus i didn't make it the default. Do you think it is too dangerous to have this option even if an admin had to implicitly add `--overwrite` to take effect of this change.

ajantha-bhat:
@MonkeyCanCode, Yes. This enables people to use same table from multiple catalogs. I don't want users to go in that direction as it is harmful for them. For example, Iceberg mentions Hadoop catalog is dangerous in production and deprecated.  Still people use it and face problems. 

So, as long as option exist, there can be cases people using it badly. 

For your benchmark, maybe drop tables (in bulk using a script) in target catalog and try register again if table states are modified.

Lastly, I really encourage people to use this catalog migrator tool and contribute more to this. Just that not enable some dangerous scenarios.

MonkeyCanCode:
> @MonkeyCanCode, Yes. This enables people to use same table from multiple catalogs. I don't want users to go in that direction as it is harmful for them. For example, Iceberg mentions Hadoop catalog is dangerous in production and deprecated. Still people use it and face problems.
> 
> So, as long as option exist, there can be cases people using it badly.
> 
> For your benchmark, maybe drop tables (in bulk using a script) in target catalog and try register again if table states are modified.
> 
> Lastly, I really encourage people to use this catalog migrator tool and contribute more to this. Just that not enable some dangerous scenarios.

I think that is fair point. I will close this PR for now.

PR #228, opened 2026-05-14
Add Polaris Tools threat model and security guidance for agents

snazy:
This adds an initial `SECURITY-THREAT-MODEL.md` and a new `AGENTS.md` for the Polaris Tools repository.

The goal is to give human reviewers and automated coding/security agents a shared description of tool-specific security boundaries, protected assets, trust boundaries, deployment responsibilities, and reporting expectations.

This is intended as analysis and triage guidance only. It does not change the project security policy, define ASF severity, decide CVE/advisory handling, or replace `SECURITY.md` or the public security reporting page.

The first version of the threat model is intentionally broad, and `AGENTS.md` is intentionally brief outside the security guidance. Follow-up PRs can refine individual tool sections if reviewers think some areas need more detail or different wording.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,145 @@
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one
+  or more contributor license agreements.  See the NOTICE file
+  distributed with this work for additional information
+  regarding copyright ownership.  The ASF licenses this file
+  to you under the Apache License, Version 2.0 (the
+  "License"); you may not use this file except in compliance
+  with the License.  You may obtain a copy of the License at
+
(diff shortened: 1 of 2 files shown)


dimas-b (commented):
I assume the text is equivalent to what we have in the main Polaris repo, right?

snazy:
> I assume the text is equivalent to what we have in the main Polaris repo, right?

It's intended to follow the main repository. The term "equivalent" is maybe a bit too strict, because each tool is very unique. If the question is whether it follow the same pattern, then yes - that's the intent.

dimas-b (approved):

Issue #231, opened 2026-05-26
Failed to start polaris-console - bind() to 0.0.0.0:80 failed (13: Permission denied)


SBatais:
Hello

I have deployed the polaris-console with the helm chart on a kubernetes cluster version 1.29.15 (managed kubernetes cluster Flexible Engine CCE), and it fails to start with the error `nginx: [emerg] bind() to 0.0.0.0:80 failed (13: Permission denied)`

I did the same on GCP GKE cluster version 1.33, and I didn't encountered the same problem.

```bash
Configured nginx for static file serving
Generated config.js with runtime configuration:
// Runtime configuration generated from environment variables
window.APP_CONFIG = {
  VITE_POLARIS_API_URL: 'https://xxxxxxxxxxxxxxxxxxxx',
  VITE_POLARIS_REALM: 'POLARIS',
  VITE_POLARIS_PRINCIPAL_SCOPE: 'openid profile email',
  VITE_OAUTH_TOKEN_URL: 'https://xxxxxxxxxxxxxxxxxx/api/catalog/v1/oauth/tokens',
  VITE_POLARIS_REALM_HEADER_NAME: '',
  VITE_OIDC_ISSUER_URL: 'https://zzzzzzz',
  VITE_OIDC_CLIENT_ID: 'yyyyyyyyyyyyyyyyyyyyyy',
  VITE_OIDC_REDIRECT_URI: 'https://xxxxxxxxxxxxxxxxxx/auth/callback',
  VITE_OIDC_SCOPE: 'openid profile email'
};
nginx: [emerg] bind() to 0.0.0.0:80 failed (13: Permission denied)
```

Thanks in advance for the help to fix this issue

jbonofre:
Thanks for the report. Let me take a look.

SBatais:
I tried to do the installation with the helm custom value `service.targetPort: 8080`, and I got the same error `nginx: [emerg] bind() to 0.0.0.0:80 failed (13: Permission denied)`. Therefore I think that the port of the container is not configurable as it's currently implemented.

MonkeyCanCode:
I think that is due to we are using non-root user to bind on port 80 which can be disallowed on certain OS. We could add this in the chart to implicitly add the grant but I think it may be better to move away from default and use something likes 8080 instead.

MonkeyCanCode:
[...]

PR #234, opened 2026-06-01
Console: chart enhance

MonkeyCanCode:
This PR polish the helm chart used by Polaris Console by matching to the same layout as well as re-using chart templates from Apache Polaris. I also updated both top level and console project level makefile to reflect those and add them to the CI.

This PR also also change the nginx port as well as k8s service pods from `80` to `8080`. This is done to avoid issue likes https://github.com/apache/polaris-tools/issues/231.

PR #236, opened 2026-06-01
Console: Change console to use non-80 default port

MonkeyCanCode:
We may want some discussions around this before merge. Currently we are using port `80` on nginx for port binding and this can be problematic as it is a privilege port and can cause issues such as https://github.com/apache/polaris-tools/issues/231.

Generally, I would change port `80` to `8080` instead. However, from our example, we also using keycloak on localhost and that is using `8080` as well. It may be confusing if we are using `8080` as well (or we should just update doc to map keycloak to other port?

### DIFF
--- a/console/README.md
+++ b/console/README.md
@@ -240,7 +240,7 @@ make build-docker
 Then, you run Polaris Console using:
 
 ```bash
-docker run -p 8080:80 \
+docker run -p 4000:4000 \
   -e VITE_POLARIS_API_URL=http://polaris:8181 \
   -e VITE_POLARIS_REALM=POLARIS \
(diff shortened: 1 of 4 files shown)


jbonofre (approved):

jbonofre:
The documentation should be updated, but I will tackle that async as I have more WIP.

MonkeyCanCode:
> The documentation should be updated, but I will tackle that async as I have more WIP.

As discussed over DM, this port had been changed to 8080 with doc updated to state keycloak will be running on port 18080. This is tracked via https://github.com/apache/polaris-tools/pull/234

PR #251, opened 2026-07-14
CI: Replace blocked Gradle action

ajantha-bhat:
Replace the blocked gradle/actions/setup-gradle action with GitHub-owned Gradle caching via actions/setup-java across all Gradle-based workflows.

Observed that PR builder was not running
https://github.com/apache/polaris-tools/actions/runs/29258683748

`The action gradle/actions/setup-gradle@v4 is not allowed in apache/polaris-tools because all actions must be from a repository owned by your enterprise, created by GitHub, or match one of the patterns: `

### DIFF
--- a/.github/workflows/apprunner-ci.yml
+++ b/.github/workflows/apprunner-ci.yml
@@ -59,15 +59,7 @@ jobs:
             21
             ${{ matrix.java-version != '21' && matrix.java-version || '' }}
           distribution: 'temurin'
-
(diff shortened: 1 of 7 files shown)


snazy (changes requested):
The action is not "blocked" per-se ; it works fine in the main repo.
The version just needs to be updated to the one used in the main repo.

ajantha-bhat:
> The action is not "blocked" per-se ; it works fine in the main repo.
The version just needs to be updated to the one used in the main repo.

Thanks. I will open a new PR for that.

PR #252, opened 2026-07-14
CI: Update setup-gradle action version

ajantha-bhat:
Observed that PR builder was not running
https://github.com/apache/polaris-tools/actions/runs/29258683748

`The action gradle/actions/setup-gradle@v4 is not allowed in apache/polaris-tools because all actions must be from a repository owned by your enterprise, created by GitHub, or match one of the patterns: `

Bumped the version to be same as polaris main.

### DIFF
--- a/benchmarks/src/gatling/scala/org/apache/polaris/benchmarks/parameters/BenchmarkConfig.scala
+++ b/benchmarks/src/gatling/scala/org/apache/polaris/benchmarks/parameters/BenchmarkConfig.scala
@@ -25,7 +25,7 @@ import scala.jdk.CollectionConverters._
 object BenchmarkConfig {
   val config: BenchmarkConfig = apply()
 
-  private def readHeaders(http: Config): Map[String, String] = {
+  private def readHeaders(http: Config): Map[String, String] =
     if (http.hasPath("headers")) {
       val headersConfig = http.getConfig("headers")
       val result = collection.mutable.Map[String, String]()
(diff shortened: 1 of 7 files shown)


benchmarks/src/gatling/scala/org/apache/polaris/benchmarks/parameters/BenchmarkConfig.scala
```diff
@@ -25,7 +25,7 @@ import scala.jdk.CollectionConverters._
 object BenchmarkConfig {
   val config: BenchmarkConfig = apply()
 
-  private def readHeaders(http: Config): Map[String, String] = {
+  private def readHeaders(http: Config): Map[String, String] =
```
ajantha-bhat, line 28: spotless was failing.

polaris-synchronizer/cli/build.gradle.kts
```diff
@@ -23,7 +23,7 @@ plugins {
   `java-library`
   `maven-publish`
   signing
-  id("com.github.johnrengelman.shadow") version "8.1.1"
+  id("com.gradleup.shadow") version "9.3.2"
```
ajantha-bhat, line 26: was using old plugin and build was failing.

ajantha-bhat:
All checks passed and PR is ready. 
cc: @snazy, @dimas-b

snazy (approved):

PR #256, opened 2026-07-21
Synchronizer: Add --skip-iceberg-content flag to speed up DR-oriented syncs

saidixith002:
Add a `--skip-iceberg-content` flag to sync-polaris that skips walking and synchronizing Iceberg namespaces and tables for each catalog. Principals, principal-roles, catalogs, catalog-roles, and grants are still synchronized as before.

Motivation: disaster recovery
------------------------------
For teams running Polaris in an active/standby (or multi-region) topology, polaris-synchronizer is a convenient way to keep the *access control and catalog structure* of a standby instance in sync with the primary so that failover is fast - principals, principal-roles, catalogs, catalog-roles, and grants all need to exist ahead of time on the standby.

However, the default sync path also opens an Iceberg REST catalog session per catalog and recursively lists every namespace and table underneath it, which is the most expensive part of the sync and scales with the number of tables rather than the number of Polaris entities. For a DR-readiness sync that may run frequently (e.g. on a schedule) purely to keep entity/grant structure current, this Iceberg walk is unnecessary overhead - the actual Iceberg table metadata/content is typically already replicated by a separate, storage-level mechanism (e.g. object storage replication) rather than by this tool.

`--skip-iceberg-content` lets operators run frequent, lightweight structural syncs for DR readiness without paying the cost of a full namespace/table walk, while still running full syncs (without the flag) when an actual data migration is needed.

Known limitation
-----------------
Grant synchronization is not filtered by grant type, so a TABLE- or NAMESPACE-scoped grant on a catalog role is still synced to the target even when `--skip-iceberg-content` prevented that table/namespace from ever being created there. Against a real Polaris server this would likely fail server-side (grant referencing an unknown resource); with the default `--halt-on-failure=false` this fails silently per-grant rather than aborting the run. This is called out explicitly and covered by a test documenting the current behavior - narrowing grant sync to skip content-scoped grants when this flag is set is left as follow-up work.

Testing
-------
Added PolarisSynchronizerSkipIcebergContentTest covering:
- Iceberg REST catalog sessions are never initialized when the flag is set, while catalog-role sync still runs.
- Original (Iceberg content synced) behavior is unchanged when the flag is not set.
- Documents that TABLE-scoped grants are still attempted even when the underlying table was never synced (see Known limitation above).

### DIFF
--- a/polaris-synchronizer/api/src/main/java/org/apache/polaris/tools/sync/polaris/PolarisSynchronizer.java
+++ b/polaris-synchronizer/api/src/main/java/org/apache/polaris/tools/sync/polaris/PolarisSynchronizer.java
@@ -63,6 +63,8 @@ public class PolarisSynchronizer {
 
   private final boolean diffOnly;
 
+  private final boolean skipIcebergContent;
+
   public PolarisSynchronizer(
       Logger clientLogger,
       boolean haltOnFailure,
@@ -70,7 +72,8 @@ public class PolarisSynchronizer {
       PolarisService source,
       PolarisService target,
       ETagManager etagManager,
-      boolean diffOnly) {
+      boolean diffOnly,
+      boolean skipIcebergContent) {
     this.clientLogger =
         clientLogger == null ? LoggerFactory.getLogger(PolarisSynchronizer.class) : clientLogger;
     this.haltOnFailure = haltOnFailure;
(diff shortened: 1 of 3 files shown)


dimas-b (approved):
Thanks for your contribution, @saidixith002 !

This looks like a reasonable option to add 👍

saidixith002:
Thanks for the review @dimas-b. I do not have access to merge the PR, please help merging it.

dimas-b:
We'll merge it soon :slightly_smiling_face: The custom in Polaris is to give at least ~2 days for major change reviews (like adding config / CLI options): https://polaris.apache.org/community/contributing-guidelines/#code-contribution-guidelines

PR #258, opened 2026-07-22
Synchronizer: Add consolidated sync report and --fail-on-error flag

saidixith002:
## Motivation

  `PolarisSynchronizer` currently reports progress purely via interleaved SLF4J log lines (`clientLogger.info`/`.error`) scattered across each of its 9 sync methods. For any
  non-trivial migration — hundreds of catalogs, namespaces, or tables — this is unreadable: there's no way to tell at a glance what succeeded, what failed, or how many entities
  of each type were touched without grepping through thousands of log lines.
  
  This is especially painful for automated usage (CI, cron-based backup/sync jobs), where the only way to detect a partial failure today is to scrape logs for `ERROR` lines.
  There's also no way to fail a script/pipeline based on sync outcome — `sync-polaris` always exits `0`, even if individual entities failed to sync.

  Additionally, since the sync is designed to be idempotent (only the diff between source and target is applied), a fully-synced re-run currently logs each skipped entity
  individually but never surfaces a summary — so a user re-running the tool sees no output at all and can't easily tell "everything is already in sync" from "something silently
  didn't run."
  
  ## What this changes

  - `sync-polaris` now prints a **consolidated synchronization report** once, at the end of every run, instead of relying on interleaved logs. It summarizes, per entity type,
  how many entities were `created`, `overwritten`, `removed`, `skipped` (already in sync), and `failed`, followed by a list of failures with their identifiers and error
  messages:
  
```
    === Synchronization Report ===
    Principal:               3 created, 0 overwritten, 0 removed, 0 skipped, 1 failed
    Catalog:                 2 created, 1 overwritten, 0 removed, 5 skipped, 0 failed
    Table:                  20 created, 0 overwritten, 0 removed, 40 skipped, 2 failed

    Failures:
      - Principal 'alice': <exception message>
      - Table 'db.orders': <exception message>
    ===============================
```
  Entity types with all-zero counts are omitted to keep the report scannable, and the `Failures:` section is omitted entirely when there are none.

  - Adds a new `--fail-on-error` flag. By default, `sync-polaris` still exits `0` regardless of individual entity failures (preserving existing behavior: failures are logged
  and the run continues to completion). With `--fail-on-error`, the command exits non-zero if the report contains any failures. This is deliberately distinct from the existing
  `--halt-on-failure`, which stops the run at the *first* failure — `--fail-on-error` lets the run finish synchronizing everything it can, then fails afterward, which is
  generally what you want for a CI/cron job (best-effort sync, but still surface that something needs attention).

  - Entities that were already in sync are now tallied under `skipped` rather than just being logged and forgotten. Without this, an idempotent re-run against an already-synced
  target prints an empty report with no `===` content in between, which reads as ambiguous ("did it even check anything?") rather than reassuring.

  ## Implementation

  - New `EntityType`, `SyncOutcome`, and `SynchronizationReport` classes (`api/.../planning/plan/`), mirroring the existing `SynchronizationPlan`/`PlannedAction` pattern
  already used elsewhere in this module rather than introducing a new abstraction style.
  - `PolarisSynchronizer` takes a `SynchronizationReport` collaborator (constructor injection, same pattern as `etagManager`) and records an outcome at every
  create/overwrite/remove/skip site across all 9 sync methods.
  - `SyncPolarisCommand` constructs the report, passes it into `PolarisSynchronizer`, prints `report.render()` via the existing console logger after the run completes, and
  returns exit code `1` when `--fail-on-error` is set and `report.hasFailures()`.
  - README updated with a sample report and documentation for `--fail-on-error`.

  ## Test plan

  - [x] `SynchronizationReportTest`: `recordSuccess`/`recordFailure` target the correct cells, `hasFailures()` toggles correctly, `render()` omits all-zero entity types and the
  empty `Failures:` section, multiple failures render correctly, and `SKIPPED` is counted and rendered without being treated as a failure
  - [x] `./gradlew :polaris-synchronizer-api:test :polaris-synchronizer-cli:test` passes
  - [x] Manually verified end-to-end against a local docker-compose Polaris source/target setup: first sync shows correct `created` counts; a second, idempotent run shows
  `skipped` counts instead of an empty report

### DIFF
--- a/polaris-synchronizer/api/src/main/java/org/apache/polaris/tools/sync/polaris/planning/plan/SynchronizationReport.java
+++ b/polaris-synchronizer/api/src/main/java/org/apache/polaris/tools/sync/polaris/planning/plan/SynchronizationReport.java
@@ -1,0 +1,140 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.polaris.tools.sync.polaris.planning.plan;
+
+import java.util.ArrayList;
+import java.util.EnumMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Accumulates the outcome of every individual entity sync attempt performed during a run, so that
+ * a single consolidated summary can be printed at the end instead of relying on interleaved logs.
+ */
+public class SynchronizationReport {
+
+  private record Failure(EntityType type, String identifier, String message) {}
+
+  private final Map<EntityType, Map<SyncOutcome, Integer>> counts;
+
+  private final List<Failure> failures;
+
+  public SynchronizationReport() {
+    this.counts = new EnumMap<>(EntityType.class);
+    for (EntityType type : EntityType.values()) {
+      Map<SyncOutcome, Integer> outcomeCounts = new EnumMap<>(SyncOutcome.class);
+      for (SyncOutcome outcome : SyncOutcome.values()) {
+        outcomeCounts.put(outcome, 0);
+      }
+      this.counts.put(type, outcomeCounts);
+    }
+    this.failures = new ArrayList<>();
+  }
+
+  /**
+   * Records that an entity was successfully synced.
+   *
+   * @param type the category of entity that was synced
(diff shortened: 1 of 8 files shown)


polaris-synchronizer/api/src/main/java/org/apache/polaris/tools/sync/polaris/planning/plan/SynchronizationReport.java
```diff
@@ -0,0 +1,140 @@
...
+   * @param type the category of entity that was synced
+   * @param outcome the outcome of the sync; must not be {@link SyncOutcome#FAILED} - use {@link
+   *     #recordFailure(EntityType, String, Exception)} for failures
+   */
+  public void recordSuccess(EntityType type, SyncOutcome outcome) {
+    if (outcome == SyncOutcome.FAILED) {
+      throw new IllegalArgumentException("Use recordFailure() to record a failed sync.");
+    }
+    Map<SyncOutcome, Integer> outcomeCounts = counts.get(type);
+    outcomeCounts.put(outcome, outcomeCounts.get(outcome) + 1);
```
dimas-b, line 62: `Map.compute()`?

dimas-b (approved):
Nice enhancement 👍 Thanks again for your contribution, @saidixith002 !

Just one minor comment.

dimas-b:
@saidixith002 : please resolve merge conflicts

saidixith002:
@dimas-b , Resolved all the conflicts. Please merge

dimas-b (approved):

PR #265, opened 2026-08-07
Catalog Migrator: Add Iceberg view migration support

ajantha-bhat:
## Summary

Adds Iceberg view migration support to the Iceberg Catalog Migrator.

The migrator now supports migrating both Iceberg tables and Iceberg views from a source catalog to a target catalog without copying data. View migration uses the new Iceberg `ViewCatalog.registerView` support added in Iceberg 1.11.0, which registers a view in the target catalog from the source view metadata location.

By default, the `migrate` command now migrates both tables and views when both source and target catalogs support Iceberg views. Existing table-only selector behavior is preserved for users who explicitly provide table selectors.

## New Options

Adds view selector options to the `migrate` command:

| Option | Description |
|---|---|
| `--view-identifiers` | Comma-separated list of view identifiers to migrate |
| `--view-identifiers-from-file` | File containing one view identifier per line |
| `--view-identifiers-regex` | Regex used to select matching view identifiers |

Selector behavior:

| Table selector | View selector | What happens |
|---|---|---|
| Not specified | Not specified | Migrates **all tables** and **all views** |
| Specified | Not specified | Migrates **selected tables only**, no views |
| Not specified | Specified | Migrates **selected views only**, no tables |
| Specified | Specified | Migrates **selected tables** and **selected views** |

## Why Views Are Supported Only With `migrate`

View support is intentionally added only to the `migrate` command, not the `register` command.

For tables, `register` can leave the table entry in both source and target catalogs. That is already risky and documented, but table cleanup can use `dropTable(..., purge=false)` when migrating so only the catalog entry is removed from the source.

Iceberg views do not have an equivalent purge-false delete operation. Dropping a view is a lightweight catalog operation, so view migration must delete the successfully migrated view from the source catalog after registering it in the target catalog. Because of that, view migration requires `migrate` semantics and is rejected for `register`.

## Notes

- View migration runs only when both source and target catalogs implement Iceberg view support.
- Table migration and view migration are tracked independently in the result summary.
- Failures for tables or views are reported separately, and failed identifiers can be retried using the existing table selector options or the new view selector options.
- Documentation and examples are updated to show table and view migration usage.

### DIFF
--- a/iceberg-catalog-migrator/README.md
+++ b/iceberg-catalog-migrator/README.md
@@ -19,10 +19,10 @@
 
 # Iceberg Catalog Migrator
  
-The Iceberg Catalog Migrator is a command-line tool that enables bulk migration of [Apache Iceberg](https://iceberg.apache.org/) tables from one [Iceberg Catalog](https://iceberg.apache.org/rest-catalog-spec/) to another without the need to copy the data. This tool works with all Iceberg Catalogs; not just Polaris.
+The Iceberg Catalog Migrator is a command-line tool that enables bulk migration of [Apache Iceberg](https://iceberg.apache.org/) tables and views from one [Iceberg Catalog](https://iceberg.apache.org/rest-catalog-spec/) to another without the need to copy the data. This tool works with all Iceberg Catalogs; not just Polaris.
 
 The migrator tool provides two operations:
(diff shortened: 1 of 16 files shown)


ajantha-bhat:
@dimas-b: Please take a look when you have time. Thanks.

dimas-b (approved):

PR #270, opened 2026-09-08
Catalog Migrator: Fix META-INF legal files and exclude Hadoop webapps

ajantha-bhat:
## Summary

- add clean Apache LICENSE and project NOTICE files to every non-shadow Catalog Migrator JAR under META-INF
- retain the curated bundle legal files in the shaded CLI JAR
- exclude unused Hadoop webapps resources from the shaded CLI JAR
- remove jQuery, jQuery UI, and DataTables entries that are no longer bundled

This addresses the 1.1.0 RC0 vote feedback about missing META-INF/LICENSE and META-INF/NOTICE files in the API artifacts and the bundled jquery.jstree.js resource.

The JAR-specific legal files follow the approach used by Apache Polaris and avoid including distribution-only Gradle wrapper licensing in Maven artifacts.

## Verification

- ./gradlew clean check -x spotlessKotlinCheck
- ./gradlew assemble
- ./gradlew spotlessKotlinGradleCheck
- verified every API, API-test, CLI, source, and Javadoc JAR contains exactly one META-INF/LICENSE and META-INF/NOTICE
- verified the shaded CLI JAR contains no webapps/** resources and its embedded legal files match BUNDLE-LICENSE and BUNDLE-NOTICE
- git diff --check

The standalone spotlessKotlinCheck task is currently blocked locally by a ktfmt Parser initialization error affecting all buildSrc Kotlin files.


iceberg-catalog-migrator/gradle/jar-licenses/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
```
ajantha-bhat, line 5: Similar to main repo https://github.com/apache/polaris/commit/0fe89ec8cefff8cb5bdd443d50d9fccd729831e6

ajantha-bhat:
@snazy and @jbonofre : Please take a look, this should address RC0 issues.

ajantha-bhat:
ping @jbonofre, @snazy

dimas-b (approved):

jbonofre:
I'm reviewing it now.

iceberg-catalog-migrator/buildSrc/src/main/kotlin/Java.kt
```diff
@@ -38,6 +39,17 @@ fun Project.configureJava() {
       attributes["Implementation-Version"] = project.version
     }
     duplicatesStrategy = DuplicatesStrategy.WARN
+
+    // Shadow JARs provide bundle-specific legal files. All other JARs use the clean project legal
+    // files, which omit distribution-only content such as the Gradle wrapper attribution.
+    if (this !is ShadowJar) {
+      from(rootProject.rootDir) {
+        include("gradle/jar-licenses/LICENSE").eachFile { path = "META-INF/$sourceName" }
```
jbonofre: I think this block copies specs also ship empty `gradle/` and `gradle/jar-licenses/` directory entries into every non-shadow JAR.

`eachFile { path = ... }` only rewrites file paths, and `includeEmptyDirs` default to `true` on `Jar` tasks, so the directories are copied.

If you check on produced jar (`./gradlew assemble` on `api`, `api-test`, `cli`), I'm pretty sure you will see something like:

```
11357  META-INF/LICENSE
  310  META-INF/NOTICE
    0  gradle/
    0  gradle/jar-licenses/
```

Even if it's not a big deal, I'm pretty sure it will be flagged in the next release, so better to fix it now.

A possible fix is to make `META-INF` the destination instead of rewriting paths:


```kotlin
if (this !is ShadowJar) {
  into("META-INF") {
    from(rootProject.file("gradle/jar-licenses")) { include("LICENSE", "NOTICE") }
  }
}
```

Adding `includeEmptyDirs = false` to both `from(...)` blocks also works if you would rather keep the current shape.

jbonofre (changes requested):

iceberg-catalog-migrator/buildSrc/src/main/kotlin/Java.kt
ajantha-bhat: good catch. I will fix it. 

I followed this pattern from main repo. So, I will fix it there also after this.

ajantha-bhat:
@jbonofre: Thanks for the review. Please take another look. Locally verified that no empty gradle folder exist now.  I will fix the same in polaris main repo also.

jbonofre (approved):

iceberg-catalog-migrator/buildSrc/src/main/kotlin/Java.kt
jbonofre: It's all good, thanks!

ajantha-bhat:
@jbonofre: 
I used codex and didn't observe it created PR on release branch instead of main. 
I can make a new RC from release branch. But we should have this fix on main branch too. 

Please approve this clean cherry-pick: https://github.com/apache/polaris-tools/pull/271

